PR #8, opened 2018-03-22
Add ETag support

mrallen1:
Here's an initial implementation of ETag support (see https://github.com/hexpm/hex_erl/issues/6) which adds the notion of a `cache_dir` and adds `etag` to the list of options.  

I thought that was a clean way to handle this issue since it seems possible (probable) that we may decide to add more options which may affect HTTP headers to/from the upstream hex service.

It isn't clear to me that **writing** a tarball into a cache and maintaining its ETag hash value is out of scope for this library, so direction about that is appreciated.


src/hex_repo.erl
```diff
@@ -164,8 +172,16 @@ get_tarball(Name, Version, Options) ->
 %% Internal functions
 %%====================================================================
 
+make_headers(Options) ->
+    lists:foldl(fun set_header/2, #{}, Options).
+
+set_header({etag, ETag}, Headers) -> maps:put(<<"If-None-Match">>, ETag, Headers);
+set_header(_Option, Headers) -> Headers.
+
 get(Client, URI) ->
```
wojtekmach: since get/2 is private and we have a good case to pass headers I think we can get rid of it and instead always pass headers explicitly; or remove this indirection altogether and call hex_http:get directly, up to you :) leaving as is is fine too.

src/hex_repo.erl
```diff
@@ -152,9 +154,15 @@ get_tarball(Name, Version) ->
 get_tarball(Name, Version, Options) ->
     Client = proplists:get_value(client, Options),
     Repo = proplists:get_value(repo, Options),
-    case get(Client, tarball_uri(Repo, Name, Version)) of
+    CacheDir = proplists:get_value(cache_dir, Options),
+
+    case get(Client, tarball_uri(Repo, Name, Version), make_headers(Options)) of
         {ok, {200, _Headers, Tarball}} ->
             {ok, Tarball};
```
wojtekmach: we also need a way to get an etag from the server, how about it returns `{ok, Tarball, [{etag, ETag}]}`? cc @ericmj

src/hex_repo.erl
```diff
@@ -201,4 +217,18 @@ decode(Signed, PublicKey, Decoder, Options) ->
...
+tarball_name(Name, Version) ->
+    Name ++ "-" ++ Version ++ ".tar".
+
+load_tarball_from_cache(_Name, _Version, undefined) ->
+    {error, no_cache_dir};
+load_tarball_from_cache(Name, Version, Path) ->
+    F = filename:join(Path, tarball_name(Name, Version)),
+    try
+        file:read_file(F) % also returns {ok, Data}
+    catch
```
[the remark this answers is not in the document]
mrallen1: OK, that's cool. The errors that will occur here are the normal POSIX ones.

src/hex_repo.erl
mrallen1: OK, sounds good.

src/hex_repo.erl
mrallen1: I think this is a good way to isolate the implementation of the headers and how conditional gets are handled from upstream things that use the library. 

On the other hand the **YAGNI** principle tugs at my mind.

I think we should leave it as is, since we already have an options proplist; we might as well keep using it that way, IMO.

src/hex_repo.erl
```diff
@@ -164,8 +172,16 @@ get_tarball(Name, Version, Options) ->
 %% Internal functions
 %%====================================================================
 
+make_headers(Options) ->
+    lists:foldl(fun set_header/2, #{}, Options).
+
+set_header({etag, ETag}, Headers) -> maps:put(<<"If-None-Match">>, ETag, Headers);
```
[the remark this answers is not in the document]
mrallen1: Should I change to raw strings? Think it doesn't matter much one way or the other, but I prefer binary strings.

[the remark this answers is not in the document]
wojtekmach: ok, let's keep binary strings. After this lands I'll update rest of the library to use binaries in other places too (as discussed in https://github.com/hexpm/hex_erl/pull/11#discussion_r178441622)

src/hex_repo.erl
wojtekmach: yup, let's keep as is, thanks!

src/hex_repo.erl
```diff
@@ -148,13 +150,21 @@ get_tarball(Name, Version) ->
...
-    case get(Client, tarball_uri(Repo, Name, Version)) of
-        {ok, {200, _Headers, Tarball}} ->
-            {ok, Tarball};
+    CacheDir = proplists:get_value(cache_dir, Options),
+
+    case get(Client, tarball_uri(Repo, Name, Version), make_headers(Options)) of
+        {ok, {200, Headers, Tarball}} ->
+            {ok, Tarball, get_headers([{<<"ETag">>, etag}], Headers)};
+        {ok, {304, _Headers, _Body}} ->
+            load_tarball_from_cache(Name, Version, CacheDir);
```
wojtekmach: I think we should have the same response shape whether we're hitting service or reading from cache, so we should return the etag here too.

Somewhat related: I think on cache miss, if `cache_dir` is set we should fetch & populate the cache so that it behaves as read-through cache and thus library is more useful out of the box (e.g. you don't have to manage writing to cache yourself). This makes me think however that we're putting to many responsibilities into one `get_tarball` function (and potentially all `get_*` functions because they all could use cache). I think we can keep going with this approach in this PR but we might reconsider it before first release.

mrallen1: Yeah, good point about conflating responsibility too much into a single function. I'll logically split out some the cache stuff into it's own (probably embarrassingly) simple module.

wojtekmach:
@mrallen1 ping :)

mrallen1:
Oh man. I have been so busy with work! I will try to work on this tomorrow. Thanks for the ping.

mrallen1:
OK, rebased. Not sure what y'all think about using the process dictionary as the place to hold the cache state, but using ETS means starting a process to own it, so this seemed easier.

mrallen1:
OK, what do you think of this? :)

wojtekmach:
@mrallen1 Thank you very much! I've moved some code around in e8d382c.

mrallen1:
Looks good. Thanks for being patient with me on this.

PR #12, opened 2018-04-03
Prefer charlists over binaries

wojtekmach:


mrallen1:
I know it's already merged but 👍 from me. Looks good. Thanks!

PR #13, opened 2018-04-19
Fix unpacking symlinks

wojtekmach:
Closes hexpm/hex#549

### DIFF
--- a/src/hex_tarball.erl
+++ b/src/hex_tarball.erl
@@ -311,8 +311,20 @@ unpack_tarball(ContentsBinary, Output) ->
     end.
 
 touch(Path) ->
+    case do_touch(Path) of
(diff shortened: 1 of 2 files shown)


src/hex_tarball.erl
```diff
@@ -311,8 +311,20 @@ unpack_tarball(ContentsBinary, Output) ->
     end.
 
 touch(Path) ->
+    case do_touch(Path) of
+        ok -> ok;
+        {error, enoent} -> touch_new(Path)
```
wojtekmach: previous Elixir code worked because `File.touch` does the same thing under the hood: https://github.com/elixir-lang/elixir/blob/v1.6.4/lib/elixir/lib/file.ex#L493

PR #16, opened 2018-04-20
Return cache hit/miss when getting tarball

wojtekmach:
@mrallen1 feedback would be appreciated!

### DIFF
--- a/src/hex_repo.erl
+++ b/src/hex_repo.erl
@@ -170,14 +170,14 @@ get_tarball(Name, Version, Options) ->
 
     case get(Client, tarball_uri(Repo, Name, Version), make_headers(Options2)) of
         {ok, {200, Headers, Tarball}} ->
-            ETag = get_headers([{<<"etag">>, etag}], Headers),
+            ReturnOpts = get_headers([{<<"etag">>, etag}], Headers),
             ok = maybe_put_cache(CacheDir, tarball_filename(Name, Version), Tarball),
-            {ok, Tarball, ETag};
+            {ok, Tarball, [{cache, miss} | ReturnOpts]};
 
         {ok, {304, Headers, _Body}} ->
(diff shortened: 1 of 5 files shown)


src/hex_repo.erl
```diff
@@ -170,14 +170,14 @@ get_tarball(Name, Version, Options) ->
...
-            ETag = get_headers([{<<"etag">>, etag}], Headers),
+            ReturnOpts = get_headers([{<<"etag">>, etag}], Headers),
             ok = maybe_put_cache(CacheDir, tarball_filename(Name, Version), Tarball),
-            {ok, Tarball, ETag};
+            {ok, Tarball, [{cache, miss} | ReturnOpts]};
 
         {ok, {304, Headers, _Body}} ->
-            ETag = get_headers([{<<"etag">>, etag}], Headers),
+            ReturnOpts = get_headers([{<<"etag">>, etag}], Headers),
             {ok, Tarball} = get_cache(CacheDir, tarball_filename(Name, Version)),
```
wojtekmach: there's an edge case: we have empty cache however we somehow got an etag so we pass that and get back 304 - but this line crashes on `enoent` then. I wonder if we should either:

1. leave as is
2. since we got cache miss make another request without if-none-match to get body, this is weird
3. before making the call check if there's a cache hit and return that so we never handle 304
4. before making the call check if there's a cache hit but still make a call expecting 304, this is probably even weirder :)

mrallen1: Suggest we catch the error and fail noisily. How frequently does this behavior happen?

wojtekmach: I only hit this when playing around with it (eg writing to different cache_dir) or when writing test for the first time (2 calls, 1st without cache dir, grab etag, make 2nd call with etag and cache dir, get crash)

wojtekmach: Thanks all for feedback, leaving as is then.

PR #18, opened 2018-04-27
Add hex_api:search

wojtekmach:


test/hex_api_tests.erl
```diff
@@ -22,3 +22,8 @@ get_user_test() ->
     {ok, User} = hex_api:get_user(<<"josevalim">>, ?OPTIONS),
     #{<<"username">> := <<"josevalim">>, <<"packages">> := _} = User,
     ok.
+
+search_test() ->
+    {ok, [Package | _]} = hex_api:search(<<"ecto">>, [{sort, downloads}, {page, 1}], ?OPTIONS),
```
wojtekmach: @ericmj I'm going back'n'forth wether params (sort=downloads etc) should be a map or proplist, and whether some known values (e.g. things we can sort on) should be atoms or binaries. wdyt?

PR #25, opened 2018-07-31
Creates a create_docs/1 function in hex_tarball

supersimple:
@ericmj @wojtekmach Can you please review this and tell me if it is correct?


wojtekmach:
Thank you!

PR #30, opened 2018-08-07
Rename options to config and put it as first argument

wojtekmach:


wojtekmach:
I still can't make up my mind, but I think I like this better 😅 wdyt?

mrallen1:
👍

PR #34, opened 2018-08-20
Upgrade gpb and generated protos

tsloughter:
I also throw in an export of the `config` type so I can use it in rebar3 function specs.


src/hex_core.erl
```diff
@@ -1,6 +1,8 @@
 -module(hex_core).
 -export([default_config/0]).
 
+-export_types([config/0]).
```
wojtekmach: 👌

PR #37, opened 2018-08-23
convert eunit tests to common test suites

tsloughter:
Because CT doesn't let you change working directories (it already does change to a per-suite directory when running and advises against writing data anywhere but a priv_dir which I have configured to be created for each test case) this is a little more of a change than you may want. So I understand if you prefer to stick with eunit and close this.

Note in `hex_tarball` I added support for the same file format `erl_tar` supports where the second element isn't the contents but instead the path to the existing on disk file, while the first element is the path within the tarball it should use.

That had to be done because the tarball suite works in `priv_dir` without changing directories to it. But it may also be useful for rebar3 package creation. It may not be needed, but I know I rely on it in `rebar3_hex` currently.


wojtekmach:
@tsloughter I've enabled `ct` on travis, could you take a look at failures?

tsloughter:
Hm, guessing travis might have ended up with a rebar3/otp-21 combo that was broken because some ct output changed.

Quick fix would be to just `wget https://github.com/erlang/rebar3/releases/download/3.6.1/rebar3`, I'll have to see how to get them to upgrade the rebar3 version to 3.6.1.

Not sure about the 17.5 test. I need to install 17.5 anyway, so I'm going to install that and see if the same familiar happens for me locally.

tsloughter:
Found this still open :( https://github.com/travis-ci/travis-ci/issues/9830

tsloughter:
I'll like to the PR too https://github.com/travis-ci/travis-cookbooks/pull/990 to hopefully get someone who can merge it to notice :)

tsloughter:
So the 17 issue is that `hex_erl_tar` uses `trim_all` in binary split call but that doesn't exist in 17.

I have a fix... but first should find out why it was calling this only when my branch is run. I'm guessing it is something with the symlink I screwed up. Still looking.

tsloughter:
Ah, I see why:

```
Like the above, if you must remain USTAR compatible, you must also ensure than paths for symbolic/hard links are no more than 100 bytes, otherwise PAX headers will be used.
```

I can make the symlink not absolute. But will also include a fix for the pax decoding.

wojtekmach:
I've pushed hex_tarball changes as separate commit master before merging this; thanks again!

PR #39, opened 2018-08-24
generate html docs from edown markdown through pandoc

tsloughter:
Requires [pandoc](https://pandoc.org/)

I include the markdown in the PR because then people can also view the docs on github.

edoc/edown blow up when using code blocks in modules with the required `pre` tags to get syntax highlighting to work, shown here in edown docs https://github.com/uwiger/edown/tree/master/doc#github-customizations, if there is a binary in the code. It may be fixable with some escaping, but then the source is ugly and it is a pain to do when adding more docs.

So for now I've added a copy of `highlight.js` that only contains Erlang highlighting.

Maybe one day I or someone will get around to fixing up Erlang doc generation to not need all the hacks :)


mrallen1:
I keep thinking "modern" doc generation is one of the biggest ugly things which shouldn't be too impossible to fix in Erlang :)

ericmj:
I don't think we should include the docs on github because they will almost always be slightly out of date.

doc/build.sh
```diff
@@ -0,0 +1,22 @@
...
+
+cd $(dirname $(realpath $0))
+
+cp ../README.md index.md
+
+make clean
+make
+
+# fix internal doc links
+sed -i 's/https:\/\/github\.com\/doc\/\(.*\)\.md/\1.html/g' *.html
```
ericmj: Is it edown that generates the github.com links? We should link to hexdocs instead.

ericmj:
I think this looks good afaict but we should not include the generated markdown files in the repo.

doc/build.sh
tsloughter: It is replacing the whole github url with just the filename.

Actually might not be needed for hex_core's docs.

tsloughter:
Ah right, I'll remove the markdown.

tsloughter:
I think it is all good now.

wojtekmach:
@tsloughter looks great. Are you able to test it on osx? A few minor things are broken (no realpath(1), a bit different sed usage). If not, that's ok, I'll fix it this week.

mrallen1:
I can test on OS X if that's helpful.

wojtekmach:
I've manually cherry-picked this as 66198bd and added 86b18ba on top to work on a Mac.

It works: https://hexdocs.pm/hex_core/index.html

Thank you so much @tsloughter!

wojtekmach:
I'd like to hide `safe_erl_term` and `hex_pb_*` from generated docs, I tried manually adding:

```
%% @private
-module(...)
```

but it either threw an error:

```
/Users/wojtek/src/oss/hex_core/src/safe_erl_term.erl, in module footer: at line 9: tag @private not allowed here.
edoc: skipping source file '/Users/wojtek/src/oss/hex_core/src/safe_erl_term.erl': {'EXIT',error}.
edoc: error in doclet 'edown_doclet': {'EXIT',error}.
```

or was ignored in case of `hex_pb_*`.

Any help/PRs would be very appreciated :)

wojtekmach:
Does Dash work with other packages which are using edoc and are hosted on hexdocs.pm?

hauleth:
Yes. I have tested with `sbroker` and everything is there. 

--
Łukasz Jan Niemier

PR #46, opened 2018-12-13
Update protos to proto3 and clean up dialyzer warnings

wojtekmach:
Another stab at https://github.com/hexpm/hex_core/pull/40

### DIFF
--- a/proto/hex_pb_names.proto
+++ b/proto/hex_pb_names.proto
@@ -1,4 +1,4 @@
-syntax = "proto2";
+syntax = "proto3";
 
 message Names {
   // All packages in the repository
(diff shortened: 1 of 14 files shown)


wojtekmach:
The fields are there but they don't show up in the diff, however they still have `required`/`optional` keywords which is incorrect. Will update.

wojtekmach:
For easier testing, I've added this: https://github.com/hexpm/hexpm/commit/2538f3d37a1ef520d3d61535bbb3d26e5c2b8b4c

I verified hex integration tests against hexpm that uses this branch of hex_core (ikr):

```
(cd ../hex && HEX_CORE_PATH=../hex_core mix test)
```

and it passes.

PR #55, opened 2019-02-14
Add and export hex_core:version/0

starbelly:
- Added version/0 to hex_core which is intended to be used for
  backwards compat work-arounds by other apps and libs.

### DIFF
--- a/src/hex_core.erl
+++ b/src/hex_core.erl
@@ -1,5 +1,8 @@
 -module(hex_core).
--export([default_config/0]).
+
+-include("hex_core.hrl").
+
+-export([default_config/0, version/0]).
 
 -export_type([config/0]).
 
@@ -48,3 +51,6 @@ default_config() ->
         repo_verify => true,
         repo_verify_origin => true
     }.
+
+version() ->
+  ?HEX_CORE_VERSION.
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you!

wojtekmach:
@starbelly since Rebar will now vendor hex_core, do you think we should still have `hex_core:version/0`?

I'm not opposed to having this function, but if primary beneficiary of it will no longer need it, I'm just wondering. Besides, I think users could do: `application.get_key(hex_core, vsn)` (but only when it's added as a dep, not when vendored!)

What do you think?

cc @ericmj

starbelly:
@wojtekmach We'll probably still need to deal with backwards compat problems in rebar3_hex. The `application.get_key(hex_core, vsn)` call returns undefined for me (`application:get_key(hex_core, vsn`). But yeah, if we can get that working that would be fine 😄

wojtekmach:
@starbelly oh, the application is probably not loaded. What happens when you do the following?

```erlang
application:load(hex_core), application:get_key(hex_core, vsn).
```

if that works, the proper fix is probably to add it to application list in `.app.src`.

starbelly:
Derp. Yeah, working now. Safe to remove hex_core:version/0.

PR #66, opened 2019-07-30
Default to undefined repo_organization when key does not exist

starbelly:
In the case where repo_organization key/value is not present in a
 repository configuration set the key to undefined.

Note that this needs to be back ported to a maintenance tag for 0.5.0 for rebar3 if this PR is accepted.


wojtekmach:
Thanks! The fix is released in v0.5.1 and v0.6.1

Issue #68, opened 2019-09-04
The built in httpc requests throw MatchErrors on connection loss


joladev:
https://github.com/hexpm/hex_core/blob/master/src/hex_http_httpc.erl#L14

If your connection is gone the left hand side match fails because you actually get `{:error, {:failed_connect, [{:to_address, {'repo.hex.pm', 443}}, {:inet, [:inet], :nxdomain}]}}`

It'd be nice if that error tuple was passed through instead of having to wrap all calls in `try` and rescuing the `MatchError`. I feel that the fact that it returns a ok tuple implies that it would also return an error tuple in this case

If this sounds reasonable I can make a PR

wojtekmach:
Yup, PR sounds great. Good catch.

ericmj:
Fixed in above PR.

PR #69, opened 2019-09-04
Pass error on loss of connection instead of throwing MatchError

joladev:
As discussed in https://github.com/hexpm/hex_core/issues/68

I immediately felt horrible at my lack of erlang skills. Also unsure what to do about tests, the http client itself doesn't seem to be tested.


wojtekmach:
Thanks! I'll cut a new release next week. 

Regarding tests, the only ones that would make sense here would be some kind of end-to-end integration ones that hit some server which we don't yet have. My long term plan is to consider using mini_repo as _a_ Hex server we can end-to-end hex_core against.

PR #82, opened 2019-12-30
hex_api: transition away from http_uri

michaelklishin:
which has been deprecated in OTP 23 [master].

See erlang/rebar32195 and erlang/rebar3#2191 for the background.

I would be happy to move the `to_list/1` function to a utility module but there doesn't seem to be one at the moment. Rebar3 already has one, so I'm not sure if it's worth introducing a new module here.

### DIFF
--- a/src/hex_api.erl
+++ b/src/hex_api.erl
@@ -31,24 +31,20 @@ put(Config, Path, Body) ->
 delete(Config, Path) ->
     request(Config, delete, Path, undefined).
 
-%% OTP 21+
--ifdef (OTP_RELEASE).
-encode_query_string(List0) ->
-    %% uri_string:compose_query/1 only accepts proplists where values are lists
-    Pairs = lists:map(fun ({K, V}) -> {to_list(K), to_list(V)}
-                          end, List0),
-    list_to_binary(uri_string:compose_query(Pairs)).
--else.
 %% @private
 encode_query_string(List) ->
-    QueryString =
-        join("&",
-            lists:map(fun ({K, V}) -> to_list(K) ++ "=" ++ to_list(V)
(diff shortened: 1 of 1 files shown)


michaelklishin:
Worth mentioning that this assumes that we have to support OTP versions < 21 because, well, Rebar3 still supports OTP 18.

If that's not the case this code can be simplified.

michaelklishin:
Any feedback on this?

starbelly:
Not a hex team member, but I can say it should support all the way back to 18. If it does not then we have to do hacks in rebar3.

wojtekmach:
I'll review this more closely this week but I think this is the way to go, thank you for the PR. And yes, we need to support old OTP versions, down to OTP 17 in fact.

michaelklishin:
Just curious, why is the obligation for support of those four-and five-year-old Erlang versions? The OTP team doesn't update even OTP 19 any more. Is there a support policy that demonstrates when support for Erlang < 21 will be dropped?

wojtekmach:
hex_core is used in hex and hex is compatible with elixir 1.x, 1.0 requires otp 17. There will definitely be future releases of hex and hex_core that will drop compatibility with old versions but nothing set in stone yet.

michaelklishin:
Any feedback on this? Warnings in Rebar currently make it impossible to build it on OTP 23. We have a Rebar PR ready and waiting for this module to be upstreamed.

wojtekmach:
@michaelklishin LGTM. I made very minor changes, just moving some code around to keep the old supporting more tied together, let me know what you think. And sorry for the delay.

michaelklishin:
@wojtekmach sure, LGTM. Thanks.

PR #88, opened 2020-03-15
Remove replace/2 in favor of publish/3

starbelly:
@supersimple This targets your branch for replace functionality. Removes replace/2 in favor of publish/3


supersimple:
Thanks @starbelly :taco:

PR #90, opened 2020-03-15
Revert to using undefined vs nil

starbelly:
About 4 months ago instances of `undefined` were swapped out for `nil`. This happened as part of commit 1155a06b and ed71e5d6. The idiom in Erlang however is `undefined`. I assume this change was unintentional, but if it indeed was intentional we can just close out this PR.

### DIFF
--- a/src/hex_api_key.erl
+++ b/src/hex_api_key.erl
@@ -36,7 +36,7 @@
 %%           <<"user_agent">> => <<"hex_core/0.5.0 (httpc) (OTP/21) (erts/10.2)">>},
 %%     <<"name">> => <<"hex_core">>,
 %%     <<"permissions">> => [#{<<"domain">> => <<"api">>,<<"resource">> => <<"read">>}],
-%%     <<"revoked_at">> => nil,
+%%     <<"revoked_at">> => undefined,
 %%     <<"updated_at">> => <<"2019-02-27T14:38:54Z">>,
 %%     <<"url">> => <<"https://hex.pm/api/keys/test">>},
 %%     }]}}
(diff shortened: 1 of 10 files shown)


wojtekmach:
Yes, we talked about using `undefined` e.g. here https://github.com/hexpm/hex_core/pull/54#issuecomment-456756492.

wojtekmach:
hmm, do you have a commit that swapped undefined with nil?

starbelly:
@wojtekmach Updated the description with a link to the commit 👍

src/hex_api_key.erl
```diff
@@ -36,7 +36,7 @@
 %%           <<"user_agent">> => <<"hex_core/0.5.0 (httpc) (OTP/21) (erts/10.2)">>},
 %%     <<"name">> => <<"hex_core">>,
 %%     <<"permissions">> => [#{<<"domain">> => <<"api">>,<<"resource">> => <<"read">>}],
-%%     <<"revoked_at">> => nil,
+%%     <<"revoked_at">> => undefined,
```
wojtekmach: is this really `undefined`? I believe the backend returns `nil` and we pass that through?

starbelly: Actually yes it comes back as nil, and if there is no translation done in hex_core from nil to undefined, then this PR seems invalid. Not terribly sure it's a good idea putting said translation in 🤔

wojtekmach:
Yeah not sure either.

starbelly:
@wojtekmach @ericmj 

 After some thought this quickly has become a non-issue for me. Indeed, `undefined` is the erlang idiom, but that doesn't mean that hex_core is solely intended for erlangers, at least I don't see it that way. If anything, I would be in more favor of standardizing on a "neutral" atom or perhaps no atom at all (i.e., a bit string). However, is any of this actually worth the effort? I'm leaning towards no. 

 It may be that it's trivial to translate `nil` to `undefined` using protobuffs (I wouldn't know) and if doesn't add complication to the setup then maybe it is worth doing. However, hex_core might find itself in a boat where some apps that use hex_core want `undefined` and some want `nil`. Maybe using mime types or a new k/v in hex_config could squash that issue, but still not sure it's worth it 🤔. It seems to me that if hex.pm returns `nil` then it returns `nil` and perhaps that should be the end of the story.  After all, it tells me what I need to know. Thoughts?

starbelly:
Closing this as a non-issue.

PR #97, opened 2020-09-16
Allow development on top of OTP 24

paulo-ferraz-oliveira:
`:get_stacktrace` will be no more at that moment!

### DIFF
--- a/src/hex_pb_names.erl
+++ b/src/hex_pb_names.erl
@@ -209,8 +209,8 @@ decode_msg_1_catch(Bin, MsgName, TrUserData) ->
 -else.
 decode_msg_1_catch(Bin, MsgName, TrUserData) ->
     try decode_msg_2_doit(MsgName, Bin, TrUserData)
-    catch Class:Reason ->
-        StackTrace = erlang:get_stacktrace(),
+    catch ?HEX_CORE_CLASS_REASON_STACKTRACE(Class, Reason, Stacktrace) ->
+        StackTrace = ?HEX_CORE_GET_STACKTRACE(Stacktrace),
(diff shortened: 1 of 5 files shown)


src/hex_pb_names.erl
```diff
@@ -209,8 +209,8 @@ decode_msg_1_catch(Bin, MsgName, TrUserData) ->
 -else.
 decode_msg_1_catch(Bin, MsgName, TrUserData) ->
     try decode_msg_2_doit(MsgName, Bin, TrUserData)
-    catch Class:Reason ->
-        StackTrace = erlang:get_stacktrace(),
+    catch ?HEX_CORE_CLASS_REASON_STACKTRACE(Class, Reason, Stacktrace) ->
```
wojtekmach, line 212: The pb files are automatically generated by the gpb plugin and we don’t want to manually override them (as a last resort we could automate applying our changes but I would really like to avoid that.) Could you send a patch to gpb instead?

paulo-ferraz-oliveira: I understand that 👍. Since I was coming from `rebar3` where they are vendored I didn't assume the same could be here (still looking into it) :D No need for a patch, I guess, since it (`gpb`) already generates OTP 24 compatible code. Just trying to figure out how to either automate it or also allow for `hex_core` to receive a flag to generate the proper code.

What do you think?

paulo-ferraz-oliveira:
Yeah, this is not required. Sorry about wasting your time. Compiled `rebar3` with Erlang/OTP 19 and tried to use it with Erlang/OTP 23.

Will re-open if a related issue comes up.

Issue #77, opened 2019-12-14
Deprecate hex_http:headers map in favour of list


wojtekmach:
HTTP spec allows duplicate header names so we should allow that.

wojtekmach:
OR, which I am very much tempted to do so, keep it as as a map of binary keys and _lists_ of binary values.

PR #105, opened 2020-10-28
Prevent dialyzer warnings on OTP 23

paulo-ferraz-oliveira:
Found these while `dialyzer`'ing `rebar3`.

Opening as a draft pull request, since I don't know when this'll be ready for review, but it's not **now**.

### DIFF
--- a/src/hex_http.erl
+++ b/src/hex_http.erl
@@ -7,7 +7,9 @@
 
 -type method() :: get | post | put | patch | delete.
 -type status() :: non_neg_integer().
+-export_type([status/0]).
 -type headers() :: #{binary() => binary()}.
+-export_type([headers/0]).
 -type body() :: {ContentType :: binary(), Body :: binary()} | undefined.
 -type adapter_config() :: map().
 
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -1,0 +1,9 @@
+case os:getenv("TARGET_ERLANG_VERSION") of
+    false -> CONFIG;
+    [] -> CONFIG;
+    TargetErlangVersionString ->
+        GpbOpts = proplists:get_value(gpb_opts, CONFIG),
+        TargetErlangVersion = {target_erlang_version, list_to_integer(TargetErlangVersionString)},
+        NewGpbOpts = lists:keystore(target_erlang_version, 1, GpbOpts, TargetErlangVersion),
+        lists:keystore(gpb_opts, 1, CONFIG, {gpb_opts, NewGpbOpts})
+end.
--- a/src/hex_api_release.erl
+++ b/src/hex_api_release.erl
@@ -106,7 +106,7 @@ publish(Config, Tarball) -> publish(Config, Tarball, []).
 %%      }}}
 %% '''
 %% @end
--spec publish(hexcore:config(), binary(), publish_params()) -> hex_api:response().
+-spec publish(hex_core:config(), binary(), publish_params()) -> hex_api:response().
 publish(Config, Tarball, Params) when is_map(Config) andalso is_binary(Tarball) andalso is_list(Params)->
     QueryString = hex_api:encode_query_string([{replace, proplists:get_value(replace, Params, false)}]),
     Path = hex_api:join_path_segments(hex_api:build_repository_path(Config, ["publish"])),
--- a/src/hex_registry.erl
+++ b/src/hex_registry.erl
@@ -14,8 +14,8 @@
 ]).
 -include_lib("public_key/include/public_key.hrl").
 
--type private_key() :: public_key:rsa_private_key() | binary().
--type public_key() :: public_key:rsa_public_key() | binary().
+-type private_key() :: #'RSAPrivateKey'{} | binary().
+-type public_key() :: #'RSAPublicKey'{} | binary().
 
 %%====================================================================
 %% API functions


paulo-ferraz-oliveira:
I'm having a hard time getting rid of these
```erlang
src/hex_pb_package.erl
1395: The pattern <X, Path> can never match since previous clauses completely covered the type <integer(),['Package' | 'Release' | 'RetirementStatus' | 'reason' | 'releases' | 'retired',...]>

src/hex_registry.erl
  83: Function sign_protobuf/2 has no local return
  85: The call hex_pb_signed:encode_msg(#{'payload':=binary(), 'signature':=binary()}, 'Signed') will never return since the success typing is (#{'payload':=binary() | maybe_improper_list(binary() | maybe_improper_list(any(),binary() | []) | byte(),binary() | [])}, 'Signed') -> binary() and the contract is ('Signed'(), atom()) -> binary()
```

I bumped `rebar3_gpb_plugin` (to latest) and updated the generated files, but it made no difference.

The first issue relates to an internal function, so it seems like a generation issue.
The second issue relates to the fact we're using `optional` and not `required` in the proto message (does the message name make sense?)

ferd:
I would ignore these for now since this is auto-generated code that is vendored in.

paulo-ferraz-oliveira:
> The next update will undo all these changes.

This Pull Request is to fix `hex_core` itself. Do you mean to say that `rebar3` will stop using those (vendored modules from `hex_core`)? Even if it does, is it the only thing using `hex_core`? (I'm unaware of these relations)

ferd:
> > The next update will undo all these changes.
> 
> This Pull Request is to fix `hex_core` itself. Do you mean to say that `rebar3` will stop using those (vendored modules from `hex_core`)? Even if it does, is it the only thing using `hex_core`? (I'm unaware of these relations)

I deleted my comment after finding I were in the wrong repo.

paulo-ferraz-oliveira:
> > > The next update will undo all these changes.
> > 
> > 
> > This Pull Request is to fix `hex_core` itself. Do you mean to say that `rebar3` will stop using those (vendored modules from `hex_core`)? Even if it does, is it the only thing using `hex_core`? (I'm unaware of these relations)
> 
> I deleted my comment after finding I were in the wrong repo.

Then I'll delete my reply 😄

paulo-ferraz-oliveira:
I opened tomas-abrahamsson/gpb#188 to get more info. and figure out if there's a better way to proceed.

paulo-ferraz-oliveira:
According to https://github.com/tomas-abrahamsson/gpb/issues/188#issuecomment-719743314, `hex_core` would benefit from changing 17 to 19 in https://github.com/hexpm/hex_core/blob/master/rebar.config#L14 (in terms of `dialyzer` -related warnings). I don't know, however, if this is desired by maintainers (e.g. there's no `minimum_otp_vsn` in `rebar.config`), since I found no indication of the support versions you're aiming for.

**Note**: there's still a potential issue there, regarding the generation of specs for maps, though, so this would also benefit from waiting for that "fix".

wojtekmach:
hex_core needs to support OTP 17. This is because it needs to be usable by Hex which in turns needs to be usable by Elixir 1.0.x. At some point Hex will no longer use old Elixir versions but not quite yet.

wojtekmach:
What are other options to suppress the warnings, is there anything we can do in `rebar.config`? E.g. strip the specs? Tell dialyzer to ignore these files?

paulo-ferraz-oliveira:
I'm not sure it's possible to tell `dialyzer` to not issue warnings on specific files (or do anything with `rebar.config`, in that sense). I know `rebar3` has an option that's _kinda for that_ (`exclude_mods` - ?) but haven't tried it with `hex_core` yet (IIRC, the last time I used it it seemed to remove the module from the analysis altogether which would mean that calls to it would provoke `undefined_function_calls`).

I don't think stripping the specs would help here either, since at least one of the errors has nothing to do with the specs themselves, but with the fact that the `dialyzer` detects certain parts of the code will never be called. Since the issue is in the vendored modules (and since you have to support OTP 17+), is there any possibility that they're generated with a different `target_erlang_version`, for example?

wojtekmach:
> is there any possibility that they're generated with a different target_erlang_version, for example?

I'm not sure. Do you have a proposal how this would work while keeping the compatibility?

I can think of two workarounds:

1. we generate two sets of modules, one compatible with OTP 17 and other with something more modern, and when using the library we somehow pick one or the other
2. each `*.pb.erl` contains both versions of the generated module guarded by an `-ifdef`. (perhaps an old compiler wouldn't accept that anyway)

neither approach seems good, it adds bloat to the package and may increase maintenance burden, so I don't think I'd pursue that.

paulo-ferraz-oliveira:
> I'm not sure. Do you have a proposal how this would work while keeping the compatibility?

Not yet, no. I'll try to find the time to cook something up. In any case, I'm specifically talking about the vendored modules, which are simply generated and included somewhere else (much like `rebar3` does). So my proposal goes to simply try and adding an input arg. to `vendor.sh` and work from there.

paulo-ferraz-oliveira:
72803c6 adds the possibility to generate different Erlang-targeted `gpb` consumables without breaking interface. This will allow vendored modules to be created for a given target version (when included in `rebar3` they'll prevent certain warnings from happening, for example).

**Note**: this would effectively conclude this pull request, since I don't think it'll be possible to reach 0-warnings given the current expected constraints.

ericmj (approved):

paulo-ferraz-oliveira:
I saw, in `rebar3`, there's an expectation for a field named `name` in `hex_core`'s `config`. Is it not there by choice (or is it an error)? I guess having `_ => _` in that map would solve the issue I found, but maybe it's too permissive for you (?)

paulo-ferraz-oliveira:
Do you think anything's still missing? Could I propose a merge and release, if all's OK with you?

src/hex_http.erl
```diff
@@ -7,7 +7,9 @@
 
 -type method() :: get | post | put | patch | delete.
 -type status() :: non_neg_integer().
+-export_type([status/0]).
 -type headers() :: #{binary() => binary()}.
+-export_type([headers/0]).
```
wojtekmach, line 12: we'll be changing this type in the future (https://github.com/hexpm/hex_core/issues/77) but it's probably ok to export it since we use it elsewhere, otherwise we get the warnings.

wojtekmach:
Thank you!

paulo-ferraz-oliveira:
Thank you. Any change this gets a release soon, so it's vendored into `rebar3`?

PR #116, opened 2021-04-30
Pin gpb version

wojtekmach:


wojtekmach:
I tried adding gpb as _dep_:

    {profiles, [
        {dev, [
            {deps, [
                {gpb, "4.3.1"}
            ]},
            {plugins, [
                {rebar3_gpb_plugin, "2.12.0"}
            ]},
            {provider_hooks, [
                {pre, [
                    {compile, {protobuf, compile}},
                    {clean, {protobuf, clean}}
                ]}
            ]}
        ]},

But it is ignored:

    ~/src/hex_core[wm-gpb]% rm src/hex_pb_names.erl
    ~/src/hex_core[wm-gpb]% rebar3 as dev compile
    ===> Fetching rebar3_gpb_plugin v2.12.0
    ===> Fetching gpb v4.10.0

Pinning the package as _plugin_ works albeit prints the following non-fatal warnings:

    % rebar3 as dev compile
    ===> Plugin gpb does not export init/1. It will not be used.
    ===> Plugin gpb does not export init/1. It will not be used.
    ===> Plugin gpb does not export init/1. It will not be used.
    ===> Verifying dependencies...
    ===> Analyzing applications...
    ===> Compiling hex_core

@tsloughter if there's a nicer way to solve this, any help would be appreciated. I'm on:

```
$ rebar3 version
rebar 3.15.0 on Erlang/OTP 24 Erts 12.0
```

PR #120, opened 2021-05-19
Add dialyzer checks to CI

paulo-ferraz-oliveira:
Stuff done:

* I tweaked `plugins`, as `gpb` is not a plugin (as is already brought in by `rebar3_gpb_plugin`)
* I regenerated the protobuf files with a more recent `gpb` (without this, `dialyzer` checks fail)

Thanks to @tomas-abrahamsson for pre-fixing `gpb`'s code generation: https://github.com/tomas-abrahamsson/gpb/issues/196#issuecomment-841638415.


.github/workflows/main.yml
```diff
@@ -15,21 +15,34 @@ jobs:
       matrix:
         pair:
           - erlang: master
+            dialyzer: true
```
paulo-ferraz-oliveira, line 18: Let me know if this is "too much"™️ (?)

.github/workflows/main.yml
```diff
@@ -15,21 +15,34 @@ jobs:
...
+            rebar3: 3.16.0
           - erlang: '24.0'
+            dialyzer: true
+            rebar3: 3.16.0
           - erlang: 23.3
+            rebar3: 3.6.1
           - erlang: 22.3
+            rebar3: 3.6.1
```
paulo-ferraz-oliveira, line 26: We could pull `3.16.0` till this one, but it won't support previous versions any more.

rebar.config
```diff
@@ -21,14 +21,23 @@
...
+            {rebar3_gpb_plugin, "2.20.1"}
         ]},
         {provider_hooks, [
             {pre, [
                 {compile, {protobuf, compile}},
                 {clean, {protobuf, clean}}
             ]}
+        ]},
+        {dialyzer, [
+            {warnings, [
```
paulo-ferraz-oliveira, line 33: If we add `underspecs` it shows more stuff. Let me know if it would be better to have it.

wojtekmach:
We don’t plan to use Dialyzer so im going to close this but thank you for the PR. If you'd like to submitted another PR that just bumps gpb, if that'd make so downstream projects don’t have warnings, that'd be appreciated. If it turns out that we keep introducing changes that cause warnings in downstream projects, we will revisit this PR.

If you submit a new PR, please make sure the gpb is versioned in rebar.config. You are right that it is not a plugin and as a matter of fact it prints warnings because it tries to be used as such. I couldnt find a better way to make sure we always use the same gpb version though so if you have a better idea thatd be appreciated too.

paulo-ferraz-oliveira:
Hm... I could've just tweaked it till it got to "just gpb" + some changes,right? 😄 Pushing forward...

paulo-ferraz-oliveira:
> gpb is versioned in rebar.config

For this to happen we have to have the dep. reference, which is probably easy to do (unless I'm missing something).

paulo-ferraz-oliveira:
Ah, I forgot to state: the motivation for this is that I was running `dialyzer` in `rebar3_hex` (which depends on this) and some warnings were being affected by it, but I understand you not wanting `dialyzer`, which is why I commented on the GitHub Actions file.

wojtekmach:
Last time I tried adding `{gpb, "x.y.z"}` to deps it seemed it was ignored but adding it to plugins worked. Maybe it's a matter of rebar version, if it works on latest then that's the way to go.

paulo-ferraz-oliveira:
> Last time I tried adding `{gpb, "x.y.z"}` to deps it seemed it was ignored but adding it to plugins worked. Maybe it's a matter of rebar version, if it works on latest then that's the way to go.

It should work on latest, sure, but then OTP 17 might not 😞 (so I'll bump `rebar3` per minor until CI passes, or just quit the effort if I can't get farther...).

Issue #62, opened 2019-05-11
Safe SSL defaults for built-in HTTP adapter


wojtekmach:
hex_core ships with a built-in httpc-based adapter but it's default options are insecure (doesn't verify server certs etc). See e.g.: https://blog.voltone.net/post/23 for more info.

We should port httpc configuration from Hex here or at the very least add some docs around this topic.

essen:
Maybe you can just use this function when it is available: https://www.erlang.org/doc/apps/inets/httpc.html#ssl_verify_host_options/1

OTP-25.1+

wojtekmach:
Good call, we can follow https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/inets when we detect we're running OTP 25 and maybe warn otherwise.

essen:
Sounds good!

PR #123, opened 2021-05-24
Warn of httpc unsafe defaults

wojtekmach:
Ref: https://github.com/hexpm/hex_core/issues/62

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,11 @@ Let's use default config for now. See "Configuration" section below for customiz
 Config = hex_core:default_config().
 ```
 
+**WARNING** The built-in httpc-based HTTP client adapter has insecure default settings
+and thus it is not recommended for usage, in particular for API write operations.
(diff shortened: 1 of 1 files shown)


wojtekmach:
@ferd if you have any feedback it'd be appreciated.

ferd:
Sounds helpful, particularly on pre-24 since the warning is now added. Coupled with the OTP one, I believe this would be useful guidance.

The tricky bit about the adapter config is that (at least in our case) it's fairly static, and the SSL Opts that might require hostname-level validation and can't work without being passed the hostname on each call. There's probably a way to tweak the adapter config on each call by messing with the repo config, but in our case it ended up simpler to just have a custom adapter do the lookup.

This is in no way an argument against the http_options addition, though. They can still be very helpful.

wojtekmach:
Agreed, a custom adapter seems to be the best solution for a varying hostname. Thanks @ferd!

ericmj (approved):

PR #124, opened 2021-05-24
Update rebar3 version on CI

wojtekmach:


ferd:
FWIW the 16.x release of rebar3 has clamped up the compatibility period to the same as the OTP team -- only 3 major versions.

So if you were keeping the window open very far back for us, that's probably not necessary anymore.

paulo-ferraz-oliveira:
I'd proposed a matrix like so: https://github.com/hexpm/hex_core/pull/120/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R17 (don't mind the `dialyzer` there, since that's what eventually got the pull request rejected).

wojtekmach:
@ferd do you happen to remember the latest rebar3 that worked on OTP 17?

wojtekmach:
@ferd never mind, all is well, just setup-beam cannot find the old versions but we know why and we'll have a fix soon.

wojtekmach:
The build on master is fixed now due to erlef/setup-beam update so I'm going to go ahead and close this. In the future we might want to use newer rebars on newer OTPs but it doesn't seem to be necessary at the moment.

PR #128, opened 2021-09-28
Adds licenses module that can be vendored

supersimple:
includes test


supersimple:
@ericmj This is step one of the licenses project in Hex. It picks up from the feedback on https://github.com/hexpm/hex/pull/886

PR #130, opened 2021-12-21
Use `rebar3_ex_doc` for generating and publishing docs

starbelly:
- removed edoc opts (done by rebar3_ex_doc)
- deleted docs.sh (obsolete)
- deleted docs.config (obsolete)


rebar.config
```diff
@@ -2,6 +2,9 @@
     debug_info,
     {platform_define, "^19", 'OTP_19'}
 ]}.
+
+{project_plugins, [rebar3_ex_doc]}.
```
starbelly, line 6: Alternatively, you can just rely on this being in maintainers global config. I always forget about that, so opt to put it in project_plugins.

starbelly:
Note CI is failing because it pull in the original rebar3_ex_doc, a re-run _should_ fix the issue.

wojtekmach:
This is fantastic, thank you!

PR #131, opened 2022-02-10
Get Repository Name

aj-foster:
This PR is related to hexpm/hex#931.

~In order to enable the modification of hex repositories without the need for a `--name` flag, this change adds `hex_registry:get_repository_name/2`.~

After further discussion, this PR makes a breaking change to the return type of several registry-reading functions to return additional information that may be useful in the future.

### DIFF
--- a/src/hex_registry.erl
+++ b/src/hex_registry.erl
@@ -80,13 +80,13 @@ encode_versions(Versions) ->
 
 %% @private
 decode_versions(Payload, no_verify) ->
-    #{packages := Packages} = hex_pb_versions:decode_msg(Payload, 'Versions'),
-    {ok, Packages};
+    #{repository := Repository, packages := Packages} = hex_pb_versions:decode_msg(Payload, 'Versions'),
+    {ok, #{repository => Repository, packages => Packages}};
 
 decode_versions(Payload, Repository) ->
     case hex_pb_versions:decode_msg(Payload, 'Versions') of
         #{repository := Repository, packages := Packages} ->
-            {ok, Packages};
+            {ok, #{repository => Repository, packages => Packages}};
         _ ->
             {error, unverified}
     end.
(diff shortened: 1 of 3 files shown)


wojtekmach:
Thank you for the PR! Let's hold off on it for a moment given this comment https://github.com/hexpm/hex/pull/931/files#r803454950

aj-foster:
@wojtekmach Not sure if you wanted to go ahead and include `repository` as a top-level key; just pushed a commit that does so. Easy to change if you'd prefer to keep only `packages` for now.

wojtekmach:
@aj-foster perfect, thank you! Could you do the same thing for functions for other registry files, that is `decode_versions` and `decode_package`?

aj-foster:
Sure thing. I went ahead and had each decoder return all of the top-level keys from each file (`repository` from versions; `repository` and `name` from packages).

wojtekmach (approved):

src/hex_registry.erl
```diff
@@ -49,13 +49,13 @@ encode_names(Names) ->
 
 %% @private
 decode_names(Payload, no_verify) ->
-    #{packages := Packages} = hex_pb_names:decode_msg(Payload, 'Names'),
-    {ok, Packages};
+    #{repository := Repository, packages := Packages} = hex_pb_names:decode_msg(Payload, 'Names'),
+    {ok, #{repository => Repository, packages => Packages}};
```
wojtekmach, line 53: hmm, actually should we just go with this?

```suggestion
    {ok, hex_pb_names:decode_msg(Payload, 'Names')}.
```

as long as it's safe, i.e. decode_msg doesn't return an error tuple or something?

in hindsight, it was a big mistake to return just the subset of the protobuf, just the packages field. With this change we're in sync with the protobuf which means if we ever add another field, we'll immediately start returning it here.

aj-foster: Went ahead and made this change in [85e5964](https://github.com/hexpm/hex_core/pull/131/commits/85e59640cd48855bbf6ef0eb17d700d6fb24ac0a) — easy to revert if necessary. I also made the similar change to `decode_versions/2` and `decode_package/3`.

ericmj (approved):

wojtekmach:
Thank you! Seems there are some transient issues on CI so I'll merge this afterwards.

wojtekmach:
Thank you!

PR #138, opened 2023-02-05
Match docs tarball error to package error

jjcarstens:
This helps the cases where the package size may be fine, but the documentation is too large. Previously, it would report as `too_big`, but that excludes how big the size is and if it was too big compressed or uncompressed.

This changes the `create_docs/2` error to match the `create/2` error reporting the reason for error and the size that was calculated


ericmj (approved):

jjcarstens:
I'll fix these tests shortly. Forgot to check those 🤦🏼

jjcarstens:
Tests updated

wojtekmach:
Thank you!

PR #140, opened 2023-05-04
Create hex_repo functions

cgerling:
This PR creates the three functions mentioned in https://github.com/hexpm/hex/issues/738, those functions are required in order to complete one of the items in that issue.

### DIFF
--- a/src/hex_repo.erl
+++ b/src/hex_repo.erl
@@ -6,7 +6,8 @@
     get_versions/1,
     get_package/2,
     get_tarball/3,
-    get_docs/3
+    get_docs/3,
+    get_public_key/1
 ]).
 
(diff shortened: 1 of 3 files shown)


wojtekmach (approved):
LGTM except I don’t think we should have get_installs as it is specific to hex.pm, ie other repositories would not have it. I think public key and docs ones make sense though.

cgerling:
In this case, should we keep the existing implementation of `get_installs/0` in `hex` using `HTTP` directly?

cgerling:
I've removed the `get_installs/1` from the PR

ericmj (approved):

wojtekmach:
Thank you!

PR #148, opened 2024-06-12
Fix spec as per current implementation

paulo-ferraz-oliveira:
This is making `rebar3_hex` fail in Dialyzer analysis.


wojtekmach:
Thank you!

paulo-ferraz-oliveira:
Thanks. Do you think it'd be Ok to cut a release from this one so we can import it next to `rebar3_hex` without workarounds?

wojtekmach:
v0.10.2 is out.

PR #150, opened 2024-06-16
Fix for `retirement_reason` spec.

### LINKED ISSUE #149: The spec for retirement_reason / retirement_params may be incorrect.
The retirement reason / params is specified as as atoms and the keys for params as atoms. However, this does not work. iirc, at one point in time atoms were allowed, but that changed, and we can see that here : https://github.com/hexpm/hexpm/blob/e1251217d56ec91cdb104a27d9dd70eb08aa0f47/lib/hexpm_web/erlang_format.ex#L18

It may be there was intention to serialize atoms to strings from the client side via protobuf, but that doesn't appear to work. Testing locally nets me an http 400. 

I'm remember all this being discussed some years ago, but the memory is fuzzy around the details now. In other words, either this is a bug in encoding within hex_code (it does not fail in hex_core to be clear) or the specs need to be adjusted to reflect reality.

paulo-ferraz-oliveira:
Closes #149.

Arguably, we could just have the code "accept" the binaries (as e.g. `rebar3_hex` is doing), and crash, since the spec is the intention that was passed, but with some backward compatibility I'm not sure the proposed solution will hurt.

This is a draft pr to follow-up on the discussion for the linked issue.


src/hex_api_release.erl
```diff
@@ -14,9 +14,9 @@
 
 -type publish_params() :: [{replace, boolean()}].
 
--type retirement_reason() :: other | invalid | security | deprecated | renamed.
+-type retirement_reason() :: binary().
```
wojtekmach, line 17: we don't use this type anymore anywhere, can you remove it while at it?

paulo-ferraz-oliveira: Sure.

paulo-ferraz-oliveira: 28145ef

wojtekmach:
Thank you!

paulo-ferraz-oliveira:
Would you be Ok to do a patch-release for this, to be able to run with an updated version next to https://github.com/erlef/rebar3_hex/pull/338? Thanks.

wojtekmach:
v0.10.3 is out!

paulo-ferraz-oliveira:
Thanks much, @wojtekmach.

PR #178, opened 2026-04-13
Introduce `hex_auth` module

maennchen:
Base Module organizing Repo / API Auth Management.

This PR is WIP. I'm exploring how it could look.

Implementation PRs:
* https://github.com/hexpm/hex/pull/1143
* https://github.com/erlang/rebar3/pull/3008
* https://github.com/erlef/rebar3_hex/pull/361

### DIFF
--- a/src/hex_api_oauth.erl
+++ b/src/hex_api_oauth.erl
@@ -146,7 +146,7 @@ device_auth_flow(Config, ClientId, Scope, PromptUser, Opts) ->
                 <<"user_code">> := UserCode,
                 <<"verification_uri_complete">> := VerificationUri,
                 <<"expires_in">> := ExpiresIn,
-                <<"interval">> := Interval
+                <<"interval">> := IntervalSeconds
             } = DeviceResponse,
             ok = PromptUser(VerificationUri, UserCode),
             OpenBrowser = proplists:get_value(open_browser, Opts, false),
@@ -154,8 +154,8 @@ device_auth_flow(Config, ClientId, Scope, PromptUser, Opts) ->
                 true -> open_browser(VerificationUri);
                 false -> ok
             end,
-            Deadline = erlang:system_time(second) + ExpiresIn,
-            poll_for_token_loop(Config, ClientId, DeviceCode, Interval, Deadline);
+            ExpiresAt = erlang:system_time(second) + ExpiresIn,
+            poll_for_token_loop(Config, ClientId, DeviceCode, IntervalSeconds, ExpiresAt);
         {ok, {Status, _, Body}} ->
             {error, {device_auth_failed, Status, Body}};
         {error, Reason} ->
@@ -163,13 +163,13 @@ device_auth_flow(Config, ClientId, Scope, PromptUser, Opts) ->
     end.
 
 %% @private
-poll_for_token_loop(Config, ClientId, DeviceCode, Interval, Deadline) ->
+poll_for_token_loop(Config, ClientId, DeviceCode, IntervalSeconds, ExpiresAt) ->
     Now = erlang:system_time(second),
-    case Now >= Deadline of
+    case Now >= ExpiresAt of
         true ->
             {error, timeout};
         false ->
-            timer:sleep(Interval * 1000),
+            timer:sleep(IntervalSeconds * 1000),
             case poll_device_token(Config, ClientId, DeviceCode) of
                 {ok, {200, _, TokenResponse}} when is_map(TokenResponse) ->
                     #{
@@ -177,17 +177,19 @@ poll_for_token_loop(Config, ClientId, DeviceCode, Interval, Deadline) ->
                         <<"expires_in">> := ExpiresIn
                     } = TokenResponse,
                     RefreshToken = maps:get(<<"refresh_token">>, TokenResponse, undefined),
-                    ExpiresAt = erlang:system_time(second) + ExpiresIn,
+                    TokenExpiresAt = erlang:system_time(second) + ExpiresIn,
                     {ok, #{
                         access_token => AccessToken,
                         refresh_token => RefreshToken,
-                        expires_at => ExpiresAt
+                        expires_at => TokenExpiresAt
                     }};
                 {ok, {400, _, #{<<"error">> := <<"authorization_pending">>}}} ->
-                    poll_for_token_loop(Config, ClientId, DeviceCode, Interval, Deadline);
+                    poll_for_token_loop(Config, ClientId, DeviceCode, IntervalSeconds, ExpiresAt);
                 {ok, {400, _, #{<<"error">> := <<"slow_down">>}}} ->
                     %% Increase polling interval as requested by server
-                    poll_for_token_loop(Config, ClientId, DeviceCode, Interval + 5, Deadline);
+                    poll_for_token_loop(
+                        Config, ClientId, DeviceCode, IntervalSeconds + 5, ExpiresAt
+                    );
                 {ok, {400, _, #{<<"error">> := <<"expired_token">>}}} ->
                     {error, timeout};
                 {ok, {Status, _, #{<<"error">> := <<"access_denied">>} = Body}} ->
--- a/src/hex_api_oauth.erl
+++ b/src/hex_api_oauth.erl
@@ -4,12 +4,29 @@
 -export([
     device_authorization/3,
     device_authorization/4,
+    device_auth_flow/4,
+    device_auth_flow/5,
     poll_device_token/3,
     refresh_token/3,
     revoke_token/3,
     client_credentials_token/4,
     client_credentials_token/5
 ]).
+
+-export_type([oauth_tokens/0, device_auth_error/0]).
+
+-type oauth_tokens() :: #{
+    access_token := binary(),
+    refresh_token => binary() | undefined,
+    expires_at := integer()
+}.
+
+-type device_auth_error() ::
+    timeout
+    | {access_denied, Status :: non_neg_integer(), Body :: term()}
+    | {device_auth_failed, Status :: non_neg_integer(), Body :: term()}
+    | {poll_failed, Status :: non_neg_integer(), Body :: term()}
+    | term().
 
 %% @doc
 %% Initiates the OAuth device authorization flow.
@@ -26,7 +43,7 @@ device_authorization(Config, ClientId, Scope) ->
 %% Returns device code, user code, and verification URIs for user authentication.
 %%
 %% Options:
-%%   * `name' - A name to identify the token (e.g., hostname of the device)
+%%   * `name' - A name to identify the token (defaults to the machine's hostname)
 %%
 %% Examples:
 %%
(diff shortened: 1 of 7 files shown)


src/hex_auth.erl
```diff
@@ -0,0 +1,808 @@
...
+    repo_auth_config/0,
+    auth_prompt_reason/0,
+    with_api_opts/0
+]).
+
+%% 5 minute buffer before expiry
+-define(EXPIRY_BUFFER_SECONDS, 300).
+
+%% Default polling interval (seconds)
+-define(DEFAULT_POLL_INTERVAL, 5).
```
ericmj: Is this unused?

src/hex_auth.erl
```diff
@@ -0,0 +1,808 @@
...
+        {ok, {200, _, DeviceResponse}} when is_map(DeviceResponse) ->
+            #{
+                <<"device_code">> := DeviceCode,
+                <<"user_code">> := UserCode,
+                <<"verification_uri_complete">> := VerificationUri,
+                <<"expires_in">> := ExpiresIn,
+                <<"interval">> := Interval
+            } = DeviceResponse,
+            io:format("Open ~ts in your browser and enter code: ~ts~n", [VerificationUri, UserCode]),
+            open_browser(VerificationUri),
```
ericmj: Do we want force users to open a browser? What if we are in an environment without browsers or the CLI only wants to provide the link.

src/hex_cli_auth.erl
```diff
@@ -0,0 +1,808 @@
...
+with_api(Callbacks, Permission, BaseConfig, Fun) ->
+    with_api(Callbacks, Permission, BaseConfig, Fun, []).
+
+%% @doc
+%% Execute a function with API authentication.
+%%
+%% Resolves credentials in this order:
+%% <ol>
+%% <li>`HEX_API_KEY' env var</li>
+%% <li>Per-repo `api_key' from config (with optional OAuth exchange for hex.pm)</li>
```
ericmj, line 181: Is this for rebar? In hex we don't have API keys per repository.

ericmj (commented):
I am not sure this fits in hex_core. It's very specific to implementing OAuth for a CLI, specifically to how Hex and rebar3 work with config resolution order and environment variables. hex_core should be more generic than that, for example you should be able to use it in web services, like hexdocs or hexdiff, and this wouldn't fit for that.

I understand we want to avoid auth logic between Hex and rebar3 if it's mostly shared, but I am not sure it fits in hex_core.

maennchen:
@ericmj I'm aware that this is very much scoped towards CLIs. Is there a reason to not include a CLI only module into `hex_core`?

I would love to be able to reuse the auth logic that is frankly quite complicated between our clients. That way we can minimize maintenance work, are able to roll out features more quickly and also ensure better compatibility.

ericmj:
> I'm aware that this is very much scoped towards CLIs. Is there a reason to not include a CLI only module into hex_core?

I think we can do that if we make that clear in the docs and maybe scoping under a `hex_cli` namespace as `hex_cli_auth`.

I am still am bit hesitant at keeping it so close to how Hex and Rebar3 works with environment variables and config resolution. Would every CLI client want to do it the same way?

maennchen:
> I think we can do that if we make that clear in the docs and maybe scoping under a hex_cli namespace as hex_cli_auth.

That sounds like a reasonable name.

> I am still am bit hesitant at keeping it so close to how Hex and Rebar3 works with environment variables and config resolution. Would every CLI client want to do it the same way?

Let's not do env variables. In the case of rebar, that would be better handled in the `rebar_hex_repos` module.

src/hex_api_oauth.erl
```diff
@@ -49,17 +65,139 @@ device_authorization(Config, ClientId, Scope) ->
...
+) -> {ok, oauth_tokens()} | {error, device_auth_error()}.
+device_auth_flow(Config, ClientId, Scope, PromptUser, Opts) ->
+    case device_authorization(Config, ClientId, Scope, Opts) of
+        {ok, {200, _, DeviceResponse}} when is_map(DeviceResponse) ->
+            #{
+                <<"device_code">> := DeviceCode,
+                <<"user_code">> := UserCode,
+                <<"verification_uri_complete">> := VerificationUri,
+                <<"expires_in">> := ExpiresIn,
+                <<"interval">> := Interval
```
wojtekmach: nitpick: maybe worth making the unit explicit throughout?

```suggestion
                <<"interval">> := IntervalSeconds
```

src/hex_api_oauth.erl
```diff
@@ -49,17 +65,139 @@ device_authorization(Config, ClientId, Scope) ->
...
+            } = DeviceResponse,
+            ok = PromptUser(VerificationUri, UserCode),
+            OpenBrowser = proplists:get_value(open_browser, Opts, false),
+            case OpenBrowser of
+                true -> open_browser(VerificationUri);
+                false -> ok
+            end,
+            Deadline = erlang:system_time(second) + ExpiresIn,
```
wojtekmach: nitpick: I couldn't find it it OTP docs so it's probably moot but I vaguely remember deadlines are usually in erlang monotonic time thing. Maybe worth changing the variable to `ExpiresAt` just in case? OK to leave it as is unless it resonates at all!

src/hex_api_oauth.erl
```diff
@@ -49,17 +65,139 @@ device_authorization(Config, ClientId, Scope) ->
...
-    Params0 = #{
-        <<"client_id">> => ClientId,
-        <<"scope">> => Scope
-    },
-    Params =
+    Name =
         case proplists:get_value(name, Opts) of
-            undefined -> Params0;
-            Name -> Params0#{<<"name">> => Name}
+            undefined -> get_hostname();
```
wojtekmach, line 71: just double-checking, why do we need to send user's hostname?

src/hex_cli_auth.erl
```diff
@@ -0,0 +1,703 @@
...
+                    error
+            end;
+        _ ->
+            error
+    end.
+
+%% @private
+%% Resolve OAuth token with global lock to prevent concurrent refresh attempts.
+resolve_oauth_token_with_context(Callbacks, Config) ->
+    global:trans(
```
wojtekmach, line 551: It sounds like a good idea to have this but on principle I think we wanna keep hex_core low-level (maybe to a fault), process-less, ets-less, etc, not have any global things whatsoever. So yeah, not 100% sure it belongs here.

src/hex_cli_auth.erl
```diff
@@ -0,0 +1,703 @@
...
+%%     `{error, {auth_error, no_credentials}}' instead of prompting.</li>
+%% <li>`oauth_open_browser' - When `true' (default), automatically opens the browser
+%%     during device auth flow. When `false', only prints the URL for the user.</li>
+%% </ul>
+%%
+%% Example:
+%% ```
+%% hex_cli_auth:with_api(Callbacks, write, Config, fun(C) ->
+%%     hex_api_release:publish(C, Tarball)
+%% end, [{optional, false}, {auth_inline, true}]).
```
wojtekmach: I'm curious about callbacks-based API and wrapping user code (here, `hex_api_release:publish/2`) in fun vs a more traditional imperative API. Are we wrapping user code in a fun because we might be doing multiple requests? Or we wanna trigger lifecycle hooks when we exit user code? (Or both?) Maybe just as an exercise worth exploring how an imperative API without callbacks would look like--or such is impossible, i.e. misses the point of having this module in the first place?

wojtekmach (commented):
This looks solid and well tested, I left some comments below.

src/hex_api_oauth.erl
maennchen: This is passed here to identify the device code reuqest:

https://github.com/hexpm/hexpm/blob/main/lib/hexpm/oauth/device_codes.ex#L33

We already do this presently: https://github.com/hexpm/hex/blob/main/lib/mix/tasks/hex.ex#L141

src/hex_cli_auth.erl
maennchen: Yes, this will possibly do multiple requests. For example to ask for an OTP token.

src/hex_cli_auth.erl
maennchen: What would you do as an alternative? `hex` has `Hex.OnceCache` to do the same and we need something like that in rebar as well.

Instead of a cache with locking, the new implementation persists all results directly. Therefore no cache is needed. I however still need a lock.

I could theoretically expose lock as a function that the implementation can provide instead of directly using `global`, but we would reimplement the same thing in both hex and rebar3.

ericmj:
One thing we have think about this approach, where we model hex_core very closely to how Hex/Rebar3 currently works.

What happens when I want to add a new feature to Hex, for example meta organizations/repos that can contain sub repos. Rebar3 might lag behind adding this feature or may not want to add it all, so now `hex_cli_auth` must diverge.

maennchen:
@ericmj I believe that we can deal with this when we are making changes as long as we consider rebar when making them. We can choose to build it in a way where there's no trouble not implementing a new feature, make it conditional behind a flag etc. Worst case, we can choose to fork the module.

src/hex_cli_auth.erl
```diff
@@ -0,0 +1,722 @@
...
+            end;
+        _ ->
+            error
+    end.
+
+%% @private
+%% Resolve OAuth token with global lock to prevent concurrent refresh attempts.
+resolve_oauth_token_with_context(Callbacks, Config) ->
+    global:trans(
+        {{?MODULE, token_refresh}, self()},
```
ericmj, line 552: Should we remove `self()`? Afaict this will only serialize for the current process.

src/hex_api_oauth.erl
```diff
@@ -199,3 +340,44 @@ revoke_token(Config, ClientId, Token) ->
...
+        case os:type() of
+            {unix, darwin} ->
+                {"open", [UrlStr]};
+            {unix, _} ->
+                {"xdg-open", [UrlStr]};
+            {win32, _} ->
+                {"cmd", ["/c", "start", "", UrlStr]}
+        end,
+    Port = open_port({spawn_executable, os:find_executable(Cmd)}, [{args, Args}]),
+    port_close(Port),
```
ericmj: Should we immediately close here? That's not how we do it in hex, we just let the OS clean it up. I am afraid this could race since `open_port` is async afaik.

src/hex_api_oauth.erl
```diff
@@ -199,3 +340,44 @@ revoke_token(Config, ClientId, Token) ->
...
+    {Cmd, Args} =
+        case os:type() of
+            {unix, darwin} ->
+                {"open", [UrlStr]};
+            {unix, _} ->
+                {"xdg-open", [UrlStr]};
+            {win32, _} ->
+                {"cmd", ["/c", "start", "", UrlStr]}
+        end,
+    Port = open_port({spawn_executable, os:find_executable(Cmd)}, [{args, Args}]),
```
ericmj: We should handle if `os:find_executable/1` cannot find an executable.

src/hex_cli_auth.erl
maennchen: My understanding of global is that we need to identify who is requesting the lock. It does not have to be a pid, but needs to be unique for each request.

https://www.erlang.org/doc/apps/kernel/global.html#t:id/0

maennchen: Sorry, I think I'm wrong. I'll write a good test and see what happens.

maennchen: It was correct, but `[]` nodes was not. I added a test specifically for it.

ericmj (approved):

src/hex_cli_auth.erl
wojtekmach: Sorry for not getting back to you earlier on this.

If it's straightforward enough for callers to wrap these particular hex core API calls in a lock I think that's preferred. However, if the lock is needed only deep down in the call stack, acquiring the lock up the stack would be too coarse and inefficient. So in that case, yeah, I'd probably consider yet another callback. But come to think of it, it's unclear we need that level of flexibility so it's probably best to leave this as is after all. Nevermind!

src/hex_cli_auth.erl
```diff
@@ -0,0 +1,722 @@
...
+-spec with_repo(
+    callbacks(), hex_core:config(), fun((hex_core:config()) -> Result), opts()
+) ->
+    Result | {error, auth_error()}
+when
+    Result :: term().
+with_repo(Callbacks, BaseConfig, Fun, Opts) ->
+    Optional = proplists:get_value(optional, Opts, true),
+    AuthInline = proplists:get_value(auth_inline, Opts, false),
+    case resolve_repo_auth(Callbacks, BaseConfig) of
```
wojtekmach: > `(Callbacks, BaseConfig)`

Just wondering if we should put callbacks under config?

Would users specify different functions for the same callback names in different call sites? (Even then, they could just modify Config appropriately before calling the function I suppose.) That said, config already has a bunch of stuff so I suppose it'd have to be under `oauth_callbacks` or similar.

I think the call sites would look something like this:

```erlang
f(#{oauth_callbacks := #{should_authenticate := Callback}} = Config) ->
  Callback(...),
```

which I think is ergonomic enough and it avoids the `call_callback` indirection. (Nothing wrong with it of course.)

wojtekmach (approved):
Looks good to me!

src/hex_cli_auth.erl
maennchen: So far in each implementation I used a fixed set of callbacks.

I did not add it to config since that has the type `hex_core:config()` and I did not want to add it globally.

maennchen: See:

* https://github.com/maennchen/rebar3/blob/c423bd4a350440dbdde282d0cdd5e04652c6e437/apps/rebar/src/rebar_hex_auth.erl#L33
* https://github.com/maennchen/hex/blob/e172f6bbc631ae1f14f18093906036c2cb2ef459/lib/hex/auth.ex#L38-L47

wojtekmach: I am ok with leaving this as is.

Just mentioning in case it resonates with anyone,

> I did not add it to config since that has the type hex_core:config() and I did not want to add it globally.

I think on the hex core side, the change would probably be something like this:

```diff
--- a/src/hex_core.erl
+++ b/src/hex_core.erl
@@ -126,7 +126,17 @@
     metadata_fields => all | [binary()],
     trusted => boolean(),
     oauth_exchange => boolean(),
-    oauth_exchange_url => binary() | undefined
+    oauth_exchange_url => binary() | undefined,
+    oauth_callbacks =>
+      undefined |
+        #{
+          get_auth_config => (binary() -> ...),
+          get_oauth_tokens => (-> ...),
+          persist_oauth_tokens => (..., ..., ..., ... -> ...),
+          prompt_otp => (binary() -> ...),
+          should_authenticate => (... -> ...),
+          get_client_id => (-> ...)
+        }
 }.

 -spec default_config() -> config().
@@ -157,5 +167,6 @@ default_config() ->
         metadata_fields => all,
         trusted => true,
         oauth_exchange => true,
-        oauth_exchange_url => undefined
+        oauth_exchange_url => undefined,
+        oauth_callbacks => undefined
     }.
```

and call sites in this PR. And then in hexpm/hex, we have:

```
$ rg mix_hex_core.default_config lib
lib/hex/repo.ex
307:      :mix_hex_core.default_config()

lib/hex/http.ex
14:      :mix_hex_core.default_config()

lib/hex/api/client.ex
6:      :mix_hex_core.default_config()

lib/hex/tar.ex
18:      :mix_hex_core.default_config()
```

so I think it'd be a good forcing function to have, perhaps, a single:

```elixir
defmodule Hex do
  def hex_core_config do
    %{:hex_core.default_config | ...}
  end
end
```

which seems like maybe a step in the right direction, having a single place to instantiate the config and thread it through the code.

maennchen: I’m coming around to the idea. I wouldn’t call them oauth_callbacks though. Possibly cli_auth_callbacks.

@ericmj What do you think?

ericmj: I like those suggestions

ericmj:
Thank you! 💜

PR #181, opened 2026-05-02
Stream metadata.config decoding in chunks and increase max size to 1024KB

ericmj:


jeregrine (approved):
Hitting this when we bump to latest Erlang. ❤️

dch:
On OTP28.5 + Elixir 1.19.5 we're hitting this limit already:

```
-define(MAX_METADATA_SIZE, 1024 * 1024).
```

I'll let you know what library & size we hit.

mitchellhenke:
https://hex.pm/packages/phosphor_icons seems to be a package that has a large metadata file that goes over the current limit and breaks. `hex_metadata.config` is about 399KB because it includes thousands of SVGs in the file list.

ericmj:
I checked all existing packages and no package hits the limit raised to 1024KB, some outliers are close but I think the fix should be on the package side in that case.

PR #187, opened 2026-05-17
Add Release published timestamp field to .proto

elfenlaid:
Part of https://github.com/hexpm/hex/issues/1113

Specifically https://github.com/hexpm/hex/issues/1113#issuecomment-4150889788

---

The idea is to replicate what `hexpm` shows for package release published timestamps in its web interface https://github.com/hexpm/hexpm/blob/78e8f9665992de71b22571c7d09fb738fffdd5f3/lib/hexpm_web/templates/package/versions.html.eex#L7

This `published_at` field is then going to be used to display a package release published dates in `mix hex.outdated` and for package cooldown checks described in this proposal https://gist.github.com/ericmj/16488f164ca2045e12f0f79a73c45031

### DIFF
--- a/proto/hex_pb_package.proto
+++ b/proto/hex_pb_package.proto
@@ -27,6 +27,8 @@ message Release {
   optional bytes outer_checksum = 5;
   // Indexes into Package.advisories for advisories affecting this release
   repeated uint32 advisory_indexes = 6;
+  // Release published timestamp
+  optional Timestamp published_at = 7;
 }
 
 message RetirementStatus {
@@ -77,4 +79,11 @@ message Dependency {
   optional string app = 4;
   // If set, the repository where the dependency is located
   optional string repository = 5;
+}
+
+// Based on google.protobuf.Timestamp
+// https://github.com/protocolbuffers/protobuf/blob/v3.15.8/src/google/protobuf/timestamp.proto#L136:L147
+message Timestamp {
+  required int64 seconds = 1;
+  required int32 nanos = 2;
 }
--- a/src/hex_pb_package.erl
+++ b/src/hex_pb_package.erl
@@ -69,7 +69,8 @@
         dependencies            => ['Dependency'()], % = 3, repeated
         retired                 => 'RetirementStatus'(), % = 4, optional
         outer_checksum          => iodata(),        % = 5, optional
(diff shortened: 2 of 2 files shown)


proto/hex_pb_package.proto
```diff
@@ -78,3 +80,10 @@ message Dependency {
   // If set, the repository where the dependency is located
   optional string repository = 5;
 }
+
+// Based on google.protobuf.Timestamp
+// https://github.com/protocolbuffers/protobuf/blob/v3.15.8/src/google/protobuf/timestamp.proto#L136:L147
+message Timestamp {
```
elfenlaid, line 86: The exact same type lives in the neighbour file https://github.com/hexpm/hex_core/blob/795c08eff38d0d02a305bede469eb8b3277ddcf7/proto/hex_pb_names.proto#L17-L24

Yet, I've decided against sharing this type as I think it would be easier to replace it with a native protobuf `Timestamp` available in the third version of protobuf.

src/hex_pb_package.erl
```diff
@@ -94,20 +95,19 @@
...
+      #{seconds                 => integer(),       % = 1, required, 64 bits
+        nanos                   => integer()        % = 2, required, 32 bits
+       }.
+
+-export_type(['Package'/0, 'Release'/0, 'RetirementStatus'/0, 'SecurityAdvisory'/0, 'Dependency'/0, 'Timestamp'/0]).
+-type '$msg_name'() :: 'Package' | 'Release' | 'RetirementStatus' | 'SecurityAdvisory' | 'Dependency' | 'Timestamp'.
+-type '$msg'() :: 'Package'() | 'Release'() | 'RetirementStatus'() | 'SecurityAdvisory'() | 'Dependency'() | 'Timestamp'().
 -export_type(['$msg_name'/0, '$msg'/0]).
 
--if(?OTP_RELEASE >= 24).
```
elfenlaid, line 102: Dear reviewers, please pay attention to this diff. 

I'm not sure why these checks are dropped and it's possibly an error.

I believe it's related to `gdb_opts` from https://github.com/hexpm/hex_core/blob/22992eb0bb61155901b88ea591ce297c5949684c/rebar.config#L7-L17 or to my Erlang version: `Erlang/OTP 27 [erts-15.2.7.4]`

ericmj:
Thank you! 💜

PR #194, opened 2026-06-07
Restructure Policy into per-repository restrictions and overrides

ericmj:
Replace the ordered Rule list with a repeated RepositoryPolicy, one entry per repository the policy constrains (in practice "hexpm" and the org's own repository). Each entry has:

  * restriction — baseline advisory_min_severity / retirement_reasons / cooldown limits applied to every release in the repository
  * overrides — per-package ALLOW/DENY with an optional requirement

For each candidate release the client matches the entry for its repository, then evaluates overrides (most specific requirement wins; ALLOW bypasses the restriction, DENY blocks), then the restriction. An ALLOW override is exempt from the restriction; everything else in the repository is subject to it. The Filter, Rule, oneof action, and Availability messages are gone.

visibility is unchanged. Regenerate hex_pb_policy.erl and update the policy test fixtures.

Also see https://github.com/hexpm/specifications/pull/62.

### DIFF
--- a/proto/hex_pb_policy.proto
+++ b/proto/hex_pb_policy.proto
@@ -1,5 +1,7 @@
 syntax = "proto2";
 
+import "hex_pb_package.proto";
+
 message Policy {
   // Name of repository
   required string repository = 1;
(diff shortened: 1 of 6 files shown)


maennchen (approved):

src/hex_core.erl
```diff
@@ -114,7 +111,6 @@
     repo_name => binary(),
     repo_public_key => binary(),
     repo_url => binary(),
-    repo_organization => binary() | undefined,
```
maennchen, line 117: That's not quite what I meant. I more or less did the opposite in the Rebar PR and use this instead of fiddling with the URI: https://github.com/erlang/rebar3/pull/3008

ericmj: I will revert this, I shouldn't have removed repo_organization.

The issue is that Hex stores full URLs so we cannot do the same approach as rebar3 without breaking already configured repos.

Will change so that hex_repo:get_policy uses hex_repo:build_url like hex_repo:get_package so we get the correct fallbacks.

maennchen (approved):

PR #197, opened 2026-06-15
Serialize device auth to reuse concurrent logins

maennchen:
## Problem

`device_auth` was the only auth-resolution path in `hex_cli_auth` not wrapped in `global:trans`. Concurrent callers that all needed to authenticate (e.g. parallel commands run with no credentials) each kicked off their own device auth flow — prompting the user and persisting a token multiple times instead of sharing a single login.

## Fix

- Wrap `maybe_authenticate_and_retry` in a single global lock. The device auth token is always persisted to the **global** OAuth scope, so the lock is intentionally *not* keyed by repo.
- Inside the lock, re-resolve auth first and reuse a token that **differs** from the one the caller arrived with — a caller that was waiting on the lock picks up the login a prior winner just persisted instead of prompting again.
- The "differs from current" check covers both entry reasons:
  - `no_credentials` — no key in config, so any resolved token differs → reuse.
  - `token_refresh_failed` — the rejected key is in config, so the winner re-runs device auth while a later waiter reuses the freshly persisted token.
  - It also avoids an infinite retry loop when the server keeps rejecting a token that still looks valid by expiry.

## Test

Adds a regression test that drives concurrent callers through a barrier and blocks the winner inside `should_authenticate`, asserting exactly one caller is prompted and the persisted login is reused by the rest. It fails against the unfixed code (multiple concurrent prompts) and passes with the fix. Full suite: 33/33.

### DIFF
--- a/src/hex_cli_auth.erl
+++ b/src/hex_cli_auth.erl
@@ -321,7 +321,38 @@ repo_name(_) ->
 
 %% @private
 %% Ask user if they want to authenticate, and if yes, initiate device auth.
+%%
+%% Serialized with a global lock so concurrent callers don't each trigger their
+%% own device auth flow. The first caller to acquire the lock runs device auth
+%% and persists the resulting token; subsequent callers re-check for an existing
+%% (now-valid) token inside the lock and reuse it instead of re-authenticating.
 maybe_authenticate_and_retry(BaseConfig, Fun, Reason, Opts) ->
+    global:trans(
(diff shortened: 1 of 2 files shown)


ericmj (approved):

ericmj:
@maennchen Let me know if this is ready to merge

maennchen:
@ericmj Ready. But I'll open more for other stuff surfacing over in `hex` later.

